PR #1, opened 2019-09-08
Implement MVVM and MVP

mihaicris:
Hi, 

This is my proposal for MVVM and MVP.

For **MVVM**, I didn't create separated viewModel for the ErrorView, since it is inside the `tableView.tableHeaderView`. The same for **MVP**, I didn't create a separated errorPresenter for the same reason. Would you agree?

I could not make smaller commits becasue added the integration test `test_loadFeedCompletion_rendersErrorMessageOnErrorUntilNextReload` forced me commit more files at once.

The only thing I was not able to achieve was to simulate (trigger) the existing tap gesture on the ErrorView label. An alternative would be to change the inner `UILabel` to a `UIButton`, to have an `UIControl` (`UIButton`) instead of `UIResponder` (`UIView`) element on which I could use existing `simulateTap()` helper function. 

If you have any comments, thank you.

Mihai

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,4 @@ os: osx
 osx_image: xcode11
 language: swift
 script: 
-  - xcodebuild clean build test -project UIDesignPatternsChallenge.xcodeproj -scheme "CI" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO -sdk iphonesimulator -destination "platform=iOS Simulator,OS=12.2,name=iPhone 8" ONLY_ACTIVE_ARCH=YES
+  - xcodebuild clean build test -project UIDesignPatternsChallenge.xcodeproj -scheme "CI" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO -sdk iphonesimulator -destination "platform=iOS Simulator,OS=13.0,name=iPhone 11 Pro Max" ONLY_ACTIVE_ARCH=YES
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: osx
-osx_image: xcode10.2
+osx_image: xcode11
 language: swift
 script: 
   - xcodebuild clean build test -project UIDesignPatternsChallenge.xcodeproj -scheme "CI" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO -sdk iphonesimulator -destination "platform=iOS Simulator,OS=12.2,name=iPhone 8" ONLY_ACTIVE_ARCH=YES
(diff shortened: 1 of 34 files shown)


caiozullo:
Hi Mihai, congratulations you've passed the challenge! 🥇

You're right, you could, but you don't need to create an `ErrorViewModel` in the MVVM solution or an `ErrorPresenter` in the MVP solution.

However, in the MVP solution, we'd recommend you to follow the `Interface Segregation Principle` and separate `Error` and `Loading` with distinct protocols and models. Something like:

```
struct FeedErrorViewModel {
    let message: String
}

protocol FeedErrorView {
}
```

This way, you allow different types/instances to handle each responsibility (even if now they're both implement in the same class).

---

Regarding the tap gestures, they're not very testable APIs. Here are a couple of ways you could test it:

1. The easiest one is to create a `UITapGestureRecognizer` subclass that exposes the target/action properties that you can use to trigger the tap in tests. Downside: It'll require changes in your production code to use this subclass.

2. Since `UITapGestureRecognizer` is an `NSObject` subclass, you could use the `value(forKey:)` APIs to access its private target/action APIs. Downsides: It uses private APIs, which can potentially change and break your tests!

3. UI Tests. You can create UI tests that load the simulator, render the view, and actually tap (hit) on the view. Downsides: It's super slow since you must run the application and wait for the animation to finish.

4. *Since this is a trivial functionality* (it doesn't impact the product/feature), you could decide not to write automated tests for it, and rely on manual checks.

Keep up the great work! ✅

mihaicris:
Thank you for the comments!

caiozullo:
Hi @mihaicris  If you updated to Xcode 11 / Swift 5.1, you need to update the `.travis.yml` config to run with the `xcode11` image ✅

mihaicris:
Hi Caio, yes! I saw this, good to have tests :)

mihaicris:
I am having some difficulties splitting into FeedView and FeedErrorView in MVP implementation. I think you had a different suggestion than this current implementation: 

```swift
protocol FeedLoadingView {
	func display(_ viewModel: FeedLoadingViewModel)
}
```
I am still thinking how to do it, but it seems that I have to change also the composer and the whole chain.

caiozullo:
@mihaicris Yes, our suggestion was to create a new `FeedErrorView` protocol. 

When you make structural design changes, you'll often have to update the composition layer. That's fine, that's why the composition layer is there. 

The composition layer is the centralized place for composition, so structural changes will not cascade throughout the system.

mihaicris:
Rewatching the episode with MVP implementation and looking again at the UML diagrams helped me fixed the issue. Waiting for the final OK :). Thank you!

caiozullo:
That's it! Great job 🥇

PR #4, opened 2020-01-26
My minimal solution to the challenge

sirEven:
Hey Caio and Mike, 

here comes my contribution to the ui pattern challenge. I was quite lost and have very little time, so I tried to follow your MVC Solution as closely as possible and checked out mihaicris' solution from last year as well.  This is also the reason for my very explicit commit messages, I used them to digest all the code changes.  

Thanks for your great course :) 
Cheers, Severin


MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -24,9 +29,15 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	@IBAction private func refresh() {
 		delegate?.didRequestFeedRefresh()
 	}
-	
+
+    func display(_ viewModel: FeedErrorViewModel) {
+        errorView?.show(message: viewModel.errorMessage)
+        refreshControl?.endRefreshing()
```
caiozullo: You don't need to call `endRefreshing` here in the `display(FeedErrorViewModel)`  as the presenter already calls `display(FeedLoadingViewModel(isLoading: false))` [in the case of an error](https://github.com/essentialdevelopercom/ios-lead-essentials-ui-design-patterns-challenge/pull/4/files#diff-4dede7b7862bdc4895fa59777939d2bfR48). 

This way, you treat UI updates in isolation and you don't need to duplicate code.

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -24,9 +29,15 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
+    }
+
 	func display(_ viewModel: FeedLoadingViewModel) {
 		if viewModel.isLoading {
+            errorView?.hideMessage()
```
caiozullo: I recommend moving this decision to the presenter layer. For example, the presenter could call `display(FeedErrorViewModel)` when the loading start. 

Then, you simplify the `display` method implementations as you won't have to deal with multiple UI updates in the same method. The `display(FeedLoadingViewModel)` will only update the `UIRefreshControl` view. And the `display(FeedErrorViewModel)` implementation will only update the `ErrorView`.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -5,67 +5,75 @@
...
+            if isLoading {
+                self?.errorView?.hideMessage()
+                self?.refreshControl?.beginRefreshing()
+            } else {
+                self?.refreshControl?.endRefreshing()
+            }
+        }
+        viewModel?.onLoadingError = { [weak self] loadingError in
+            self?.errorView?.show(message: loadingError)
+        }
```
caiozullo: Similar to the MVP comments, I recommend updating only the UIRefreshControl view within the `onLoadingStateChange` closure. And perform updates to the ErrorView only in the `onLoadingError` closure. It's easier to understand and maintain isolated UI updates.

So, you can move the decision to update the error state to the ViewModel. For example, the ViewModel could have an `onLoadingErrorStateChange` closure with an optional error message. If the controller receives an error message, you show the `ErrorView`. Otherwise, you hide it.

sirEven:
Hey Caio, 
Thanks a lot for your in depth answer! I will try to make the according changes as soon as possible. 

Greetings, 
Severin

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
sirEven: /removed the line.

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
sirEven: /changed the according to your input

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
sirEven: /changed the according to your input

sirEven:
do I need to close the old pull request in order to make a new one?

caiozullo:
That's it @sirEven! No need to create a new PR. 

Great job ✅

sirEven:
Thanks :)

PR #7, opened 2020-04-16
UI Design Patterns Tests

H2wk:
Hi

I enjoyed the project but still struggle a bit with MVP. Will need to review the course work again. 

I have left a few comments on my code.


MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,6 +280,13 @@ final class FeedUIIntegrationTests: XCTestCase {
 		}
 		wait(for: [exp], timeout: 1.0)
 	}
+    
+    func test_loadFeedCompletion_correctlyRendersErrorState() {
+        let(sut, _) = makeSUT()
+        
+        sut.loadViewIfNeeded()
+        assertErrorMessageIsExpected(sut: sut, expectedMessage: nil)
```
H2wk, line 288: I was unsure if my function `assertErrorMessageIsExpected ` should take in a `comment` string for example as when this method is full written out its a bit unclear as to what step in the process you are at.

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -308,4 +308,8 @@ final class FeedUIIntegrationTests: XCTestCase {
 	private func anyImageData() -> Data {
 		return UIImage.make(withColor: .red).pngData()!
 	}
+    
+    private func assertErrorMessageIsExpected(sut: FeedViewController, expectedMessage: String?, file: StaticString = #file, line: UInt = #line) {
```
H2wk, line 331: I went a bit back and forth with if the function `assertErrorMessageIsExpected ` should take in a `comment` string to provide additional context to what is expected. When this method has more steps in it, it a bit unclear as to what you're expecting after running the code.

MVP Module/Source/Feed UI/Views/ErrorView.swift
```diff
@@ -30,7 +30,7 @@ public final class ErrorView: UIView {
 		}
 	}
 	
-	@IBAction func hideMessage() {
+	@IBAction public func hideMessage() {
```
H2wk, line 33: `IBActions` are internal by default

MVVM Module/Source/Feed UI/Views/ErrorView.swift
```diff
@@ -30,7 +30,7 @@ public final class ErrorView: UIView {
 		}
 	}
 	
-	@IBAction func hideMessage() {
+	@IBAction public func hideMessage() {
```
H2wk, line 33: `IBActions` are internal by default

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
caiozullo: Go ahead - a comment is great when you need to provide additional context to explain the failure.

MVVM Module/Source/Feed UI/Views/ErrorView.swift
caiozullo: I believe you're exposing the method as a trade-off for the sake of testing - that's fine. You could have used @testable too. 

Or a much safer choice in this case: replace the UIView+TapGesture with a UIButton, which gives you two main benefits: better UX/accessibility and an easier and more reliable way to simulate taps during tests.

MVP Module/Source/Feed UI/Views/ErrorView.swift
caiozullo: I believe you're exposing the method as a trade-off for the sake of testing - that's fine. You could have used @testable too. 

Or a much safer choice in this case: replace the UIView+TapGesture with a UIButton, which gives you two main benefits: better UX/accessibility and an easier and more reliable way to simulate taps during tests.

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
caiozullo: Go ahead - a comment is great when you need to provide additional context to explain the failure.

PR #11, opened 2020-05-06
Challenge implementation

TulioOParreiras:
Feature implemented in MVVM and MVP:
- Display localized error message on feed load failure;
- Dismiss error view on tap action


MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -31,6 +41,7 @@ final class FeedPresenter {
 
 	func didStartLoadingFeed() {
 		loadingView.display(FeedLoadingViewModel(isLoading: true))
+        errorView.display(FeedErrorViewModel(showError: false))
```
caiozullo, line 44: Suggestion: Instead of passing a boolean, you could pass the error message `String`. So, the view controller is not responsible for deciding which string to use (this allows you to change error messages dynamically without changing the UI layer).

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,13 +13,23 @@ protocol FeedView {
...
+    private let errorView: FeedErrorView
 	
-	init(feedView: FeedView, loadingView: FeedLoadingView) {
+	init(feedView: FeedView, loadingView: FeedLoadingView, errorView: FeedErrorView) {
 		self.feedView = feedView
 		self.loadingView = loadingView
+        self.errorView = errorView
 	}
 
 	static var title: String {
```
caiozullo, line 35: I noticed you moved the localized strings to a `Localized` type (that's a good refactoring!). You could then remove this `title` static var from the FeedPresenter.

MVP Module/Source/Feed UI/Composers/FeedUIComposer.swift
```diff
@@ -30,7 +31,7 @@ public final class FeedUIComposer {
 		let storyboard = UIStoryboard(name: "Feed", bundle: bundle)
 		let feedController = storyboard.instantiateInitialViewController() as! FeedViewController
 		feedController.delegate = delegate
-		feedController.title = title
+		feedController.title = Localized.feedTitle
```
caiozullo, line 34: This method receives a `title` param, but you're ignoring it in favor of the `Localized.feedTitle` - you could rather pass the `Localized.feedTitle` as the param to this method call.

PR #15, opened 2020-07-12
iCamilo - UI design patterns challenge

iCamilo:
### On this PR

✅ Show an error message to the user when the load feed fails
✅ Implement it for both, MVP and MVVM UI design patterns
✅ Localize the displayed message
✅ Include an additional language: Spanish(Latinamerica)
✅ Add tap to dismiss gestures to the error indicator

Challenge | challenge2


MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -28,10 +30,15 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
 			refreshControl?.beginRefreshing()
+            errorView.hideMessage()
 		} else {
 			refreshControl?.endRefreshing()
 		}
 	}
+    
+    func display(_ viewModel: FeedErrorViewModel) {
+        errorView.show(message: viewModel.message)
+    }
```
iCamilo, line 41: Dunno why the formatting looks off in the PR. It looks ok on my local XCode 

Screen Shot 2020-07-11 at 11 10 58 PM

MVVM Module/Source/Feed UI/Views/ErrorView.xib
```diff
@@ -48,6 +48,7 @@
                     </subviews>
```
iCamilo, line 48: I used a tap gesture for the tap to dismiss configured directly from the StoryBoard. 
No tests were added, as I don't know how to simulate them. From your responses in the lectures and other PRs, I understand that as this is not a feature that in any way affects the correct behaviour of the app, it will be ok to be tested manually. 

The approaches that I've found in internet are cumbersome and they do not look "clean" to me. 
https://medium.com/@borutt/unit-testing-gesture-recognizers-b55444884864
This one for example, mocks the view controller in order to remove the actual gesture to replace it by a subclass that is testable :S 

Of course, I will rather back the implementation with tests, if you could show me how, particularly for the case where all the configuration is done in the SB 

👽🙇🏼‍♂️ Thanks!

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -43,6 +43,14 @@ extension FeedViewController {
 	var isShowingLoadingIndicator: Bool {
 		return refreshControl?.isRefreshing == true
 	}
+    
+    var isShowingErrorIndicator: Bool {
+        return errorView.message != nil
+    }
```
iCamilo, line 49: As I didn't see as correct to turn into public the `ErrorView` `isVisible` method for the sake of the tests, the DSL is deriving this value from this logic on the `ErrorView`

```
public var message: String? {
		get { return isVisible ? label.text : nil }
	}
``` 

It looks to me as a potential issue that could affect the tests if it changes for whatever reason.
Thoughts?

Thanks!

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,7 +280,61 @@ final class FeedUIIntegrationTests: XCTestCase {
...
+    
+    func test_errorIndicator_hidesAsSoonAsUserInitiatiesReloadAndBeforeItCompletes() {
+        let (sut, loader) = makeSUT()
+        sut.loadViewIfNeeded()
+        loader.completeFeedLoadingWithError()
+        XCTAssertTrue(sut.isShowingErrorIndicator, "Error indicator should be shown as feed load didFail")
+        
+        sut.simulateUserInitiatedFeedReload()
+        XCTAssertFalse(sut.isShowingErrorIndicator, "Error indicator should be hidden as a reload is initiated and not wait until load completes")
+    }
```
iCamilo, line 325: This test was added after running the app for a manual revision. I noticed that the error indicator was not dismissed once the refresh started and it was causing a weird user experience, even though the behavior was correct.
I decided to leave this test here as a way of documenting the finding, but someone may argue that is not necessary as the previous tests are somehow validating the same. 

Thoughts? 

Thanks!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -26,19 +30,24 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
+        viewModel?.onFeedLoadFails = { [weak self] error in
+            self?.errorView.show(message: (self?.viewModel?.loadingFeedErrorMessage)!)
+        }
 	}
 	
+    public override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        tableView.sizeTableHeaderToFit()
+    }
```
iCamilo, line 49: This logic here is not having the expected effect, to accommodate the table header depending on if error indicator is displayed or not. I think that is because I am using a `.xib` to create the `ErrorView` instead of layering it directly as part of the table view controller, but I am not sure. 

After expending sometime debugging, I still not understand why it is not working. 

- could you please give me some idea of what do I need to review
- is this "bug" something that deserves a test to replicate and validate the solution? If so, what should I test, that the height is cero when the error view is not displayed?

Thanks!

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
caiozullo: It's ok to add more tests than needed as you develop a feature. As you progress, I recommend you review and remove the tests that become irrelevant.

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,7 +280,61 @@ final class FeedUIIntegrationTests: XCTestCase {
 		}
 		wait(for: [exp], timeout: 1.0)
 	}
-	
+    
+    func test_feedViewLoad_doesNotShowErrorIndicatorAtFeedLoadSucceeds() {
```
caiozullo, line 284: You're missing a test for the expected initial state (not showing error):

```
let (sut, loader) = makeSUT()
		
sut.loadViewIfNeeded()

XCTAssertFalse(sut.isShowingErrorIndicator, ...)
```

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
caiozullo: Alternatively, you could have just used the `errorMessage` helper directly in all tests:

```
XCTAssertEqual(sut.errorMessage, nil)
```

Or:

```
XCTAssertEqual(sut.errorMessage, expectedMessage)
```

This way, you also increase coverage since you guarantee the error message is correct in all states.

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,6 +280,59 @@ final class FeedUIIntegrationTests: XCTestCase {
 		}
 		wait(for: [exp], timeout: 1.0)
 	}
+    
+    func test_feedViewLoad_doesNotShowErrorIndicatorAtFeedLoadSucceeds() {
```
caiozullo, line 284: You're missing a test for the expected initial state (not showing error):

```
let (sut, loader) = makeSUT()
		
sut.loadViewIfNeeded()

XCTAssertFalse(sut.isShowingErrorIndicator, ...)
```

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -28,10 +30,15 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	func display(_ viewModel: FeedLoadingViewModel) {
 		if viewModel.isLoading {
 			refreshControl?.beginRefreshing()
+            errorView.hideMessage()
```
caiozullo, line 33: The `Presenter` could decide when you hide the message. So you can control every change to the `errorView` in a centralized place:

```
func display(_ viewModel: FeedErrorViewModel) {
    if let message = viewModel.errorMessage {
        errorView.show(message: viewModel.message)
    } else {
        errorView.hideMessage()
    }
}
```

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
iCamilo: Done!

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
iCamilo: Done!

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
iCamilo: Done!

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
iCamilo: Got it and thanks!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
iCamilo: Done!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -26,19 +30,24 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
 		}
-	}
-
-	public override func viewDidLayoutSubviews() {
-		super.viewDidLayoutSubviews()
-		
-		tableView.sizeTableHeaderToFit()
+                        
+        viewModel?.onFeedLoadFails = { [weak self] error in
+            self?.errorView.show(message: (self?.viewModel?.loadingFeedErrorMessage)!)
```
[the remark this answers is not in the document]
iCamilo: Done!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -26,19 +30,24 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	func bind() {
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+                self?.errorView.hideMessage()
```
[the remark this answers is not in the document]
iCamilo: Done and thanks!

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
iCamilo: Done and thanks!

MVVM Module/Source/Feed UI/Views/ErrorView.xib
iCamilo: I did it for the MVP module. One thing I noticed is that by using a button I can`t pull to refresh if I start the pull on top of the `ErrorView` which for me is not a good UX. 
That doesn't happen with the tap gesture. 

In this case I consider that UX is more important that testability. What do you think?

iCamilo:
@caiozullo thanks for the feedback!

All the comments have been addressed. Please review at your convenience 

One comment regarding replacing tap gesture for a button:

I did it for the MVP module and I noticed that by using a button I can`t pull to refresh if I start the pull on top of the `ErrorView` which for me is not a good UX.  That doesn't happen with the tap gesture. 

In this case, I consider that UX is more important that testability. What do you think?

Thanks!

MVP Module/Source/Feed UI/Views/ErrorView.swift
```diff
@@ -6,11 +6,10 @@ import UIKit
 
 public final class ErrorView: UIView {
     @IBOutlet private var container: UIView!
-	@IBOutlet private var label: UILabel!
-    @IBOutlet public private(set) var dismiss: UIButton!
+    @IBOutlet public private(set) var label: UIButton!
```
iCamilo, line 9: Hi @caiozullo here I have to make the outlet public private(set) for the sake of being able to simulate a tap from the tests. Is this ok ?

caiozullo: Yes, that's ok. Alternatively, you can it `internal` and use `@testable import` in the test side.

caiozullo:
@iCamilo 

Regarding the `UIButton` canceling the `UITableView` scrolling, this the desired touch UX.

Remember that `UITableView` is a `UIScrollView`. And since you have the `canCancelContentTouches` and `delaysContentTouches` set to `true`, you can still scroll freely on top of the button unless you press the button too long.

A long press shows intent to interact with the button. While a quick press and scroll show intent to interact with the scroll view.

This is desired in touch devices. It may be hard to replicate this in the simulator, so I recommend trying it on a real device. ✅

PR #16, opened 2020-07-26
MVVM and MVP implementation

spromeu:
Hi guys,

This is my proposal for MVVM and MVP error message implementation.
I decided do not create separated ViewModel and Presenter for ErrorView, but I separated Loading and Error protocols and models for MVP solution because I implemented RefreshControl and ErrorView into distinct classes.

Thanks,
Sarah

### DIFF
--- a/MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
+++ b/MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
@@ -59,6 +59,9 @@ final class FeedUIIntegrationTests: XCTestCase {
         sut.simulateUserInitiatedFeedReload()
         loader.completeFeedLoadingWithError(at: 1)
         XCTAssertEqual(sut.errorMessage, "Couldn't connect to server", "Expected error message once user initiated loading completes with error")
+        
+        sut.simulateUserInitiatedFeedReload()
+        XCTAssertNil(sut.errorMessage, "Expected no error message once user initiates a reload")
 
     }
 
--- a/MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
+++ b/MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
@@ -48,10 +48,13 @@ final class FeedUIIntegrationTests: XCTestCase {
 	}
     
     func test_loadFeedCompletion_displaysErrorMessageOnError() {
-        let (sut, _) = makeSUT()
+        let (sut, loader) = makeSUT()
         
         sut.loadViewIfNeeded()
         XCTAssertNil(sut.errorMessage, "Expected no error message once view is loaded")
+        
+        loader.completeFeedLoading(at: 0)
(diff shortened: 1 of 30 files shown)


MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -46,6 +46,24 @@ final class FeedUIIntegrationTests: XCTestCase {
...
+        loader.completeFeedLoading(at: 0)
+        XCTAssertNil(sut.errorMessage, "Expected no error message once loading completes successfully")
+        
+        sut.simulateUserInitiatedFeedReload()
+        loader.completeFeedLoadingWithError(at: 1)
+        XCTAssertEqual(sut.errorMessage, "Couldn't connect to server", "Expected error message once user initiated loading completes with error")
+        
+        sut.simulateUserInitiatedFeedReload()
+        XCTAssertNil(sut.errorMessage, "Expected no error message once user initiates a reload")
+
```
caiozullo, line 65:
```suggestion
```

caiozullo:
Hi @spromeu, thank you for the PR. 

We updated the challenge repo recently, but you still had the old version. So there were a lot of conflicts in this PR. 

That's not a problem at all - I just changed the base branch from `master` to `challenge-v1` (old version) and pushed a small commit to trigger CI again.

Everything looks good now. I'll review your solution soon. ✅

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -46,6 +46,23 @@ final class FeedUIIntegrationTests: XCTestCase {
...
+        
+        sut.loadViewIfNeeded()
+        XCTAssertNil(sut.errorMessage, "Expected no error message once view is loaded")
+        
+        loader.completeFeedLoading(at: 0)
+        XCTAssertNil(sut.errorMessage, "Expected no error message once loading completes successfully")
+        
+        sut.simulateUserInitiatedFeedReload()
+        loader.completeFeedLoadingWithError(at: 1)
+        XCTAssertEqual(sut.errorMessage, "Couldn't connect to server", "Expected error message once user initiated loading completes with error")
```
caiozullo, line 61: This test seems fragile. It could fail if running in a different locale (e.g., Portuguese) because it's using the hard-coded English string. You could use the [localized test helper method](https://github.com/essentialdevelopercom/ios-lead-essentials-ui-design-patterns-challenge/blob/master/MVP%20Module/Tests/Feed%20UI/Helpers/FeedUIIntegrationTests%2BLocalization.swift) to check by the localized `key` instead of the `value`: 

```suggestion
        XCTAssertEqual(sut.errorMessage, localized("FEED_VIEW_CONNECTION_ERROR"), "Expected error message once user initiated loading completes with error")
```

spromeu:
Hi Caio,

My bad. I forgot to refactor the test to use the key after I created the localized strings file.

Many thanks for your feedback

PR #17, opened 2020-09-13
Error messages - MVVM and MVP implementation.

AlfredoHernandez:
- Displays an error message to the customer when the app fails to load the feed.
- The error message is localized in at least 4 languages.
- The customer is able to tap to dismiss the error message.
- The layout matches the MVC guide and supports Dark Mode.


MVP Module/Tests/Feed UI/FeedUISnapshotTests.swift
```diff
@@ -50,8 +43,40 @@ class FeedUISnapshotTests: XCTestCase {
...
+                controller: self,
+                imageLoader: loader
+            ),
+            loadingView: self,
+            feedErrorView: self
+        )
+        adapter.presenter = presenter
+        presenter.errorMessage = errorMessage
+        delegate = adapter
+        simulateUserInitiatedFeedReload()
```
caiozullo, line 80: In this MVP implementation, the ViewController is decoupled from the Presenter and the FeedLoader.

One of the advantages of this approach is that you don't need a `Presenter`,  an `Adapter`, or a `FeedLoader` to test the ViewController! 

You can call `display(FeedErrorViewModel)` directly to display the error message:

```
		display(FeedErrorViewModel(message: errorMessage))
    }
}
```

As simple as that! 

So you can remove the `FeedLoaderStub`.

It's important to note that this is also possible in the MVVM or MVC approach, as long as you decouple the FeedLoader from the ViewController and the ViewModel.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -17,16 +17,22 @@ final class FeedViewModel {
 	var title: String {
         Localized.Feed.title
 	}
+    
+    var errorMessage = Localized.Feed.loadError
```
caiozullo, line 21: There's no need for a mutable `var` since the `errorMessage` never changes. You can turn it into a computed property instead.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -17,16 +17,22 @@ final class FeedViewModel {
...
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+    var onFeedFailed: Observer<String>?
```
caiozullo, line 25: As you did in the MVP solution, the error message could be an `Optional<String>` - so the ViewModel can send `.none` when the UI should hide the error. You can rename it as well to make this behavior clear:

```
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -17,16 +17,22 @@ final class FeedViewModel {
...
 	
 	func loadFeed() {
 		onLoadingStateChange?(true)
+        let errorMessage = self.errorMessage
```
caiozullo, line 29: Following my previous comment, before starting to load the feed, you can send an error state change event to hide the error:

```
onErrorStateChange?(.none)
```

So the UI state transitions are all centralized in the ViewModel. The UI simply follows the event changes.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -28,6 +30,7 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
 				self?.refreshControl?.beginRefreshing()
+                self?.errorController?.hideError()
```
caiozullo, line 33: This logic to hide the error when loading the feed should be in the ViewModel. See my comments regarding the `onErrorStateChange` events. So you could remove this line and make the `hideError` method private in the error controller.

MVVM Module/Tests/Feed UI/FeedUISnapshotTests.swift
```diff
@@ -66,7 +59,12 @@ private class FeedLoaderStub: FeedLoader {
 
 private extension FeedViewController {
     func display(errorMessage: String) {
-        fatalError("Must be implemented - follow the MVC solution as a guide")
+        let error = NSError(domain: "any", code: 0)
+        let loader = FeedLoaderStub(.failure(error))
+        viewModel = FeedViewModel(feedLoader: loader)
+        errorController?.viewModel = viewModel
+        errorController?.viewModel?.errorMessage = errorMessage
+        simulateUserInitiatedFeedReload()
```
caiozullo, line 67: To simulate an error, you can just call the `onFeedFailed` (or `onErrorStateChange`). And since you're using an error controller in your implementation, you can share the same ViewModel:

```
        errorController?.viewModel = viewModel
        viewModel?.onFeedFailed?(errorMessage)
    }
}
```

caiozullo (commented):
Hi @AlfredoHernandez,

- All tests are passing ✅
- Added an extra language ✅

I left a couple of comments/suggestions. Feel free to update the code based on the review and push it again.

Great job!

PR #28, opened 2020-11-28
Robert Dates - UI Design Patterns Challenge

rsndates:


MVP Module/Source/Feed Presentation/FeedLoadingViewModel.swift
```diff
@@ -4,4 +4,5 @@
 
 struct FeedLoadingViewModel {
 	let isLoading: Bool
+    let error: String?
```
caiozullo, line 7: I recommend separating the `error` from the `loading` view model in this case, as they can be used by different views / controllers.

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -23,15 +23,15 @@ final class FeedPresenter {
...
+
 	func didFinishLoadingFeed(with feed: [FeedImage]) {
 		feedView.display(FeedViewModel(feed: feed))
-		loadingView.display(FeedLoadingViewModel(isLoading: false))
+        loadingView.display(FeedLoadingViewModel(isLoading: false, error: nil))
 	}
 	
 	func didFinishLoadingFeed(with error: Error) {
-		loadingView.display(FeedLoadingViewModel(isLoading: false))
+        loadingView.display(FeedLoadingViewModel(isLoading: false, error: Localized.Feed.loadError))
```
caiozullo, line 35: I recommend separating the `errorView` from the `loadingView` in this case, as they can be implemented by different views / controllers.

You could create a new `ErrorView` protocol, for example.

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -24,19 +24,31 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
-	
+    
+  private func showErrorMessage(message: String?) {
+      if let message = message {
+          errorView?.show(message: message)
+      } else {
+          errorView?.hideMessage()
+      }
+      refreshControl?.endRefreshing()
+  }
```
caiozullo, line 46: Showing an error and showing a loading indicator should be separate concerns.

For example, it could show an error while loading - and it could *not* show an error while not loading. And so on.

The presentation logic is already defined by the `Presenter`, so the UI shouldn't decide when to show/hide the error depending on the loading state with the chained if/else statements. 

It'd be best to render the views based on the data received. "If there's an error string, show the error regardless of the loading state." "If isLoading is true, show the loading indicator regardless of the error message state." And so on.

This way, the views and controllers are much easier to implement and understand. No need for chained if/else.

caiozullo (commented):
Hi @rsndates,

- All tests are passing ✅
- Added an extra language ✅

I left some suggestions. Feel free to push changes.

Great job!

PR #43, opened 2021-01-26
Ricardo Herrera - UI design Patterns Challenge

calimat:
Hi @caiozullo and @mapostolakis:  

I am still not finished with this, I only started the MVVM tests, and I am still stuck in the snapshot tests. 

I see that it works for the light mode snapshot test but it doesn't work with the dark snapshot test. The image created by the tests on the MVVM module is a an image with a white line in the top of the device which in the stored files that image does not exist. I tried hiding the view, changing the background color to clear color, and also I used the different simulators iPhone 12 iOS 14.2 on Xcode 12.3 and still no luck.

I know that the images don't match but I still don't know how to fix this. Any ideas ?


caiozullo:
Hi @calimat, you need to update the table view in the storyboard to adapt to dark mode. 

You need to set the background color to `System Background Color` for all views - including the table footer and header view! The `System Background Color` will update automatically for dark/light mode.

You can change the storyboard setting to show the view in dark mode, so it's easier to see which images you need to update.

calimat:
Hi @caiozullo Thank you. I solved it with the system background color. 

I committed all MVVM and MVP. Please review and let me know if have completed the challenge successfully or there are other changes I should make. 

Thanks again.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,17 @@ final class FeedViewModel {
 	
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedError: Observer<String?>?
```
caiozullo, line 23: `onFeedError` could be called `onErrorStateChange` since it also accepts a *no error case* (`nil`)

caiozullo (commented):
Hi @calimat,

- All tests are passing ✅
- Added an extra language ✅

I left only one review comment. Feel free to push changes.

Great job!

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
calimat: Got it, thanks.

caiozullo:
Nice one!

PR #45, opened 2021-02-01
Adrian Szymanowski - UI Design Patterns Challenge 

Adrimi:


MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,16 @@ final class FeedViewModel {
 	
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadWithError: Observer<Error>?
```
caiozullo, line 23: `onFeedLoadWithError: Observer<String> ` could be `onErrorStateChange: Observer<String?>`, so the ViewModel decides when to show/hide errors (instead of the view controller). It's easier to manage state when all changes are centralized in a single component.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -20,6 +21,7 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	}
 	
 	@IBAction private func refresh() {
+		errorView?.hideMessage()
```
caiozullo, line 24: The ViewModel should decide when to show/hide messages. See my other comment.

caiozullo (commented):
Hi @Adrimi,

- All tests are passing ✅
- Added an extra language ✅

Great job!

Adrimi:
Thank you @caiozullo for the suggestions! I totally missed the point that `FeedViewController` was changing the error state by directly calling `errorView?.hideMessage()`

Also, the onErrorStateChange: Observer<String?> appears to be simpler form than observing `Error` type. In that challenge there is no complicated error changing, so simple String works just fine :)

caiozullo:
Nice one!

PR #54, opened 2021-02-22
Bogdan Poplauschi - UI Design Patterns Challenge

bpoplauschi:
# Intent
Implement localized user-facing error messages using TDD.

# Implementation
- updated the MVVM stack so it can display the load error message, following all the required use cases
  - added `FeedViewModel.onFeedLoadFailed` observable that exposes potential error messages
- updated the MVP stack so it can display the load error message, following all the required use cases
  - defined new abstractions `FeedErrorViewModel` and `FeedErrorView` to allow a clear separation of the new error displaying functionality
  - `FeedViewController` conforms to `FeedErrorView` - this is the MVP recommended approach, where the presenter `FeedPresenter` references the view `FeedViewController` indirectly via an abstraction - in this case `FeedErrorView`
  - make sure we don't create a retain cycle between the view and the presenter by using `WeakRefVirtualProxy`
- Dark mode support
- added Romanian localization

# Class Diagram MVC
MVC

# Class Diagram MVVM
MVVM

# Class Diagram MVP
MVP


MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,16 +18,25 @@ final class FeedViewModel {
 		Localized.Feed.title
 	}
 	
+	var errorMessage: String {
+		Localized.Feed.loadError
+	}
```
caiozullo, line 23: This property can be private. Or you can even inline it at the call site: `self.onFeedLoadFailed?(Localized.Feed.loadError)`

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,16 +18,25 @@ final class FeedViewModel {
...
+	
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadFailed: Observer<String>?
```
caiozullo, line 27: `onFeedLoadFailed: Observer<String> ` could be `onErrorStateChange: Observer<String?>`, so the ViewModel decides which message to show and when to show/hide errors (instead of the view controller). It's often easier to manage state when all changes are centralized.

For example, in the `loadFeed()` method, the ViewModel can call  `onErrorStateChange(.none)` to hide the error before loading the feed. And when there's an error: `onErrorStateChange(Localized.Feed.errorMessage)`.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -28,10 +31,14 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
 				self?.refreshControl?.beginRefreshing()
+				self?.errorView.hideMessage()
```
caiozullo, line 34: The ViewModel should decide when to hide/show errors. See my other comment.

caiozullo (commented):
Hi @bpoplauschi,

- All tests are passing ✅
- Added an extra language ✅

Great job!

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
bpoplauschi: Good catch - it's inlined now

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
bpoplauschi: This is a very good point, subtle difference, but important. I have refactored as you suggested.

bpoplauschi:
Thanks, @caiozullo!

caiozullo:
Nice one!

PR #67, opened 2021-03-20
Tak Maz - UI Design Patterns Challenge

Guy783:


MVP Module/Source/Feed Presentation/FeedErrorViewModel.swift
```diff
@@ -0,0 +1,11 @@
+//
+//  FeedErrorViewModel.swift
+//  MVPTests
+//
+//  Created by Tak Mazarura on 18/03/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+struct FeedErrorViewModel {
+	let hasError: Bool
```
caiozullo, line 10: Choosing which message to show is a presentation decision - it should be done by the presenter. 

So the `message ` could be an optional `String?` instead of a boolean. This way, Presenter can decide which message to show and when to show/hide the message.

For example, in the `didStartLoadingFeed()` method, the Presenter can call  `feedErrorView.display(FeedErrorViewModel(message: .none))` to hide the error.

And in `didFinishLoadingFeed(with error:)`, the Presenter can call  `feedErrorView.display(FeedErrorViewModel(message: Localized.Feed.loadError))` to show the error.

This way, the ViewController doesn't need to decide which error to show and when to hide the error. It's often easier to manage state when all changes are centralized.

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,25 +13,34 @@ protocol FeedView {
...
 	}
 	
 	func didStartLoadingFeed() {
 		loadingView.display(FeedLoadingViewModel(isLoading: true))
+		feedErrorView.display(FeedErrorViewModel(hasError: false))
 	}
 	
 	func didFinishLoadingFeed(with feed: [FeedImage]) {
 		feedView.display(FeedViewModel(feed: feed))
+		feedErrorView.display(FeedErrorViewModel(hasError: false))
```
caiozullo, line 38: The error is hidden on `didStartLoadingFeed` so you shouldn't do it again here. You can remove this line.

```suggestion
```

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,25 +13,34 @@ protocol FeedView {
...
 }
 
+protocol FeedErrorView {
+	func display(_ viewModel: FeedErrorViewModel)
+}
+
 final class FeedPresenter {
 	private let feedView: FeedView
 	private let loadingView: FeedLoadingView
+	private let feedErrorView: FeedErrorView
```
caiozullo, line 23: `feedErrorView` could be renamed to `errorView` since the context is clear within this class.

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -299,3 +346,38 @@ final class FeedUIIntegrationTests: XCTestCase {
 		return UIImage.make(withColor: .red).pngData()!
 	}
 }
+
+// MARK: - Test Helpers
+extension FeedUIIntegrationTests {
+	private func test_errorView(_ errorView: ErrorView?,
```
caiozullo, line 352: This method preform assertions, so it should be called 'assert' instead of 'test_errorView':

```suggestion
	private func assert(_ errorView: ErrorView?,
```

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,17 +280,64 @@ final class FeedUIIntegrationTests: XCTestCase {
 		}
 		wait(for: [exp], timeout: 1.0)
 	}
+}
+
+// MARK: - Error View Tests
+extension FeedUIIntegrationTests {
```
caiozullo, line 286: None of the test classes in this codebase use extensions with `MARK`s - so this implementation is not following the codebase convention. Please, follow the convention to make the code consistent and easier to review. Thank you!

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,17 +280,64 @@ final class FeedUIIntegrationTests: XCTestCase {
...
+		test_errorView(sut.errorView, isVisible: false)
+	}
+	
+	func test_onFeedLoadError_errorMessageShown() {
+		let (sut, loader) = makeSUT()
+		
+		sut.loadViewIfNeeded()
+		
+		loader.completeFeedLoadingWithError()
+		test_errorView(sut.errorView, isVisible: true, withMessage: "Couldn't connect to server")
```
caiozullo, line 310: This test will fail if the localized strings change. You should use the `localized` helpers as it's done in other tests using localized strings and as shown in the lectures.

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,17 +280,64 @@ final class FeedUIIntegrationTests: XCTestCase {
...
+		
+		sut.loadViewIfNeeded()
+		loader.completeFeedLoadingWithError()
+		
+		sut.errorView?.simulateTapToHideErrorView()
+		expectation(for: .init { _, _ in
+			sut.errorView?.isVisible == false
+		}, evaluatedWith: nil, handler: nil)
+
+		waitForExpectations(timeout: 2.0, handler: nil)
```
caiozullo, line 324: You shouldn't use expectations since the tap action is synchronous. Please, follow the guidelines shown in the lectures. Thank you!

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,17 +280,64 @@ final class FeedUIIntegrationTests: XCTestCase {
...
+	func test_onFeedLoadSuccess_errorMessageNotShown() {
+		let (sut, loader) = makeSUT()
+		
+		sut.loadViewIfNeeded()
+		
+		loader.completeFeedLoading(at: 0)
+		test_errorView(sut.errorView, isVisible: false)
+	}
+	
+	func test_onFeedLoadError_errorMessageShown() {
```
caiozullo, line 304: Missing test/assertion to check that the error view is hidden when the user manually reloads (pull to refresh) after an error.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,18 @@ final class FeedViewModel {
 	
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadError: Observer<Error?>?
```
caiozullo, line 23: Choosing which message to show is a presentation decision - it should be done by the view model.

`onFeedLoadError: Observer<Error>?` could be `onErrorStateChange: Observer<String?>?`, so the ViewModel decides which error to show and when to show/hide errors (instead of the view controller). It's often easier to manage state when all changes are centralized.

And on failure, the ViewModel can call  `onErrorStateChange(Localized.Feed.loadError)` to show the error.

This way, the ViewController doesn't need to decide which error to show/hide.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,18 @@ final class FeedViewModel {
...
 	
 	func loadFeed() {
 		onLoadingStateChange?(true)
 		feedLoader.load { [weak self] result in
-			if let feed = try? result.get() {
+			switch result {
+			case .success(let feed):
 				self?.onFeedLoad?(feed)
+				self?.onFeedLoadError?(nil)
```
caiozullo, line 31: `self?.onFeedLoadError?(nil)` should be done before calling `feedLoader.load`

This means there's a missing test.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -26,10 +28,20 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	func bind() {
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
+			guard let self = self else { return }
```
caiozullo, line 31: The `onLoadingStateChange` shouldn't be changed. Moreover, the `guard` is not needed - undo this change.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -26,10 +28,20 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
+			}
+		}
+		
+		viewModel?.onFeedLoadError = { [weak self] error in
+			guard let self = self else { return }
+			if error == nil {
+				self.errorView.hideMessage()
 			} else {
-				self?.refreshControl?.endRefreshing()
+				self.errorView.show(message: Localized.Feed.loadError)
```
caiozullo, line 44: The ViewModel should decide when to hide/show errors. See my other comment.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -26,10 +28,20 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
 			if isLoading {
-				self?.refreshControl?.beginRefreshing()
+				self.refreshControl?.beginRefreshing()
+			} else {
+				self.refreshControl?.endRefreshing()
+			}
+		}
+		
+		viewModel?.onFeedLoadError = { [weak self] error in
+			guard let self = self else { return }
```
caiozullo, line 40: The guard is not needed - remove it.

MVVM Module/Tests/Feed UI/FeedUIIntegrationTests.swift
```diff
@@ -280,17 +280,64 @@ final class FeedUIIntegrationTests: XCTestCase {
 		}
 		wait(for: [exp], timeout: 1.0)
 	}
+}
+
```
caiozullo, line 284: Please, apply the same improvements I suggested in the MVP solution here in the MVVM tests as well. Thank you!

caiozullo (commented):
Hi @Guy783, you almost passed the challenge!

Please, check my review comments and let me know once you're done. Thank you 🙂

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
Guy783: Updated :)

MVP Module/Tests/Feed UI/FeedUIIntegrationTests.swift
Guy783: Updated

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
Guy783: Should the `onLoadingStateChange ` not be changed, because it was not part of the challenge?

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
Guy783: Updated :) 

I do have a question though. My thinking came from a previous lecture. We do not want the following code to be executed if `self` is `nil`. Self is `weak`. Why would we not want use a `guard let self ...`in this instance.

MVP Module/Source/Feed Presentation/FeedPresenter.swift
Guy783: Updated! :)

MVP Module/Source/Feed Presentation/FeedPresenter.swift
Guy783: Updated! :)

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
caiozullo: Hi @Guy783, yes, the `onLoadingStateChange` shouldn't be changed because it's not part of the challenge.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
caiozullo: Because the scope is very short in this closure and using `self?` has the same effect - if `self` is `nil`, nothing will happen ✅

MVP Module/Source/Feed Presentation/FeedErrorViewModel.swift
Guy783: Ahh, ok, so the `ViewModel` in the MVP is all data and no behaviour, in addition the data should be in a state where it is as ready for presentation as possible such that the View does not need to have any logic for presentation as done in my previous implementation.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
Guy783: Updated! :)

MVP Module/Source/Feed Presentation/FeedErrorViewModel.swift
Guy783: Updated! :)

caiozullo:
Hi @Guy783, is this PR ready for review?

PR #91, opened 2021-06-13
Fei Wen - UI Design Patterns Challenge

fnytrip:


caiozullo:
Hi @fnytrip,

Our CI doesn't support Xcode 12.5 and iOS 14.5 yet. But there are no breaking changes from Xcode 12.4 and iOS 14.4 to Xcode 12.5 and iOS 14.5 - they produce exactly the same snapshot. So it should work just fine if you're using the right setup described in the readme file. We tested this extensively ✅

The only issue is if you're developing on an M1 because it may generate different snapshots. But we fixed this last week by adding pixel tolerance to the snapshot comparison logic. Please, make sure you're using the latest `xcode12_5` branch (you may need to [update your fork](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork)).

To pass the challenge, the baseline images and the MVC target files shouldn't have been changed. Please, update your fork and undo all changes to existing snapshots and the MVC target.

Let me know once you're done or if you have any questions/issues. 

Thanks 🙂

MVP Module/Source/Feed Presentation/FeedErrorViewModel.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  File.swift
+//  MVP
+//
+//  Created by Fei Wen on 6/12/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+struct FeedErrorViewModel {
+	let isError: Bool
```
caiozullo, line 10: The `isError: Bool` could be an optional string `message: String?`, so the Presenter can decide when to show/hide the message and which message to show.

```suggestion
	let message: String?
```

So, for example, in the `didStartLoadingFeed()` method, the Presenter can call  `feedErrorView.display(FeedErrorViewModel(message: .none))` to hide the error.

And on error, the Presenter can call  `feedErrorView.display(FeedErrorViewModel(message: Localized.Feed.loadError))` to show the formatted error.

This way, the ViewController doesn't need to check a boolean value and then show or hide the message, and the Presenter can decide to show different messages without having to alter the view controller or the view model. It's easier to manage presentation state transitions when all changes are centralized in the presenter.

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,25 +13,34 @@ protocol FeedView {
...
 }
 
+protocol FeedErrorView {
+	func display(_ viewModel: FeedErrorViewModel)
+}
+
 final class FeedPresenter {
 	private let feedView: FeedView
 	private let loadingView: FeedLoadingView
+	private let feedErrorView: FeedErrorView
```
caiozullo, line 23: `feedErrorView` could be named just `errorView` to simplify - the context is clear here so we can avoid repeating "feed" many times (like the `loadingView` is not called `feedLoadingView`).

```suggestion
	private let errorView: FeedErrorView
```

If you follow this suggestion, please, rename it everywhere in this class.

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,25 +13,34 @@ protocol FeedView {
...
+		self.feedErrorView = feedErrorView
 	}
 
 	func didStartLoadingFeed() {
+		feedErrorView.display(FeedErrorViewModel(isError: false))
 		loadingView.display(FeedLoadingViewModel(isLoading: true))
 	}
 
 	func didFinishLoadingFeed(with feed: [FeedImage]) {
+		feedErrorView.display(FeedErrorViewModel(isError: false))
```
caiozullo, line 37: Since the error is hidden on `didStartLoadingFeed`, we don't need to hide it again in this `didFinishLoadingFeed(with feed:)` method. You can remove this line and the tests still pass:

```suggestion
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,22 @@ final class FeedViewModel {
 		Localized.Feed.title
 	}
 
+	var loadError: String {
+		Localized.Feed.loadError
+	}
+
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onError: Observer<Bool>?
```
caiozullo, line 27: Just like in the MVP solution, the `var onError: Observer<Bool>?` could be defined as `var onErrorStateChange: Observer<String?>?`, so the ViewModel can decide when to show/hide the message and which message to show.

```suggestion
	var onErrorStateChange: Observer<String?>?
```

So, for example, in the `loadFeed()` method, the ViewModel can call  `onErrorStateChange?(.none)` to hide the error before loading the feed.

And on error, the View Model can call  `onErrorStateChange?(Localized.Feed.loadError)` to show the formatted error.

This way, the ViewController doesn't need to check a boolean value and then show or hide the message, and the View Model can decide to show different messages without having to alter the view controller. It's easier to manage presentation state transitions when all changes are centralized in the view model.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -32,6 +34,13 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 				self?.refreshControl?.endRefreshing()
 			}
 		}
+		viewModel?.onError = { [weak self] isError in
+			if isError {
+				self?.errorView?.show(message: self?.viewModel?.loadError ?? "")
```
caiozullo, line 39: An empty string is not a valid path `?? ""`. So the code should never display an empty error string to the user (or it'll show an empty red label).

If you follow my suggestions and move the `loadError` decision to the View Model, we won't have to handle this invalid path because the view controller will receive the error message in the closure instead of a boolean ✅

caiozullo (commented):
Hi @fnytrip, you almost passed the challenge!

Thank you 🙂

fnytrip:
Hi @caiozullo Thanks a lot for the detailed feedback. I've updated the PR accordingly.

caiozullo (commented):
Hi @fnytrip,

- All tests are passing ✅
- Added an extra language ✅

Great job!

PR #95, opened 2021-06-17
Sam Yang - UI Design Patterns Challenge

startover205:


MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,22 @@ final class FeedViewModel {
 		Localized.Feed.title
 	}
 
+	var errorMessage: String {
+		Localized.Feed.loadError
+	}
```
thiagopenna, line 23: Following from my previous comment, the way the `errorMessage` is implemented doesn't allow for different messages to be displayed, and in the end the `FeedViewController` is the one deciding which message to show. I suggest observing as optional `String?` instead.
```suggestion
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -32,6 +36,14 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 				self?.refreshControl?.endRefreshing()
 			}
 		}
+		viewModel?.onLoadingErrorOccur = { [weak self] errorOccurred in
+			guard let self = self, let viewModel = self.viewModel else { return }
+			if errorOccurred {
+				self.errorView?.show(message: viewModel.errorMessage)
```
thiagopenna, line 42: Following from the previous comments, in here the `FeedViewController` decides on which message to show based on `errorOccurred`, but the viewModel should be the one to decide on which message to display during the state change.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -23,6 +23,10 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		viewModel?.loadFeed()
 	}
 
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
```
thiagopenna, line 28: I recommend creating an IBOutlet instead of extracting the error view from the `tableHeaderView`.

```suggestion
	@IBOutlet private(set) var errorView: ErrorView?
```

thiagopenna (commented):
Hi @startover205,

This is Penna, I am an instructor at Essential Developer.

- All tests are passing ✅
- Added an extra language ✅

Great job!

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,22 @@ final class FeedViewModel {
...
+
 	var onLoadingStateChange: Observer<Bool>?
+	var onLoadingErrorOccur: Observer<Bool>?
```
thiagopenna, line 26: `onLoadingErrorOccur: Observer<Bool>?` could be `onErrorStateChange: Observer<String?>?`, so the ViewModel decides which message to show and when to hide errors (instead of the view controller). It's easier to manage presentation state transitions when all changes are centralized.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
thiagopenna, line 23: Following from my previous comment, the way the `errorMessage` is implemented doesn't allow for different messages to be displayed, and in the end the `FeedViewController` is the one deciding which message to show. I suggest observing as optional `String?` instead.
```suggestion
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
thiagopenna, line 42: Following from the previous comments, in here the `FeedViewController` decides on which message to show based on `errorOccurred`, but the viewModel should be the one to decide on which message to display during the state change.

startover205:
Hi Penna!

Thanks for the great advice! It helps me a lot.

thiagopenna:
Well done!

PR #96, opened 2021-06-17
Sanel Zsivics - UI Design Patterns Challenge

sanelz:


sanelz:
Note: it says that I was working from the xcode12_4 branch, but the screenshot below says that it's up to date with the xcode12_5 branch. I was working with Xcode 12.5 and with the iPhone 12 - iOS 14.5 simulator.


Screenshot 2021-06-17 at 22 07 27

caiozullo:
Hi @sanelz, you probably started the challenge before we created the `xcode12_5` branch. If you have no conflicts, that's fine.

But please, undo all changes to the schemes (it shouldn't have been changed - please check the readme for more details). Let me know once you're done so we can start the review. Thank you!

sanelz:
Hi @caiozullo, yes I started the challenge before `xcode12_5` was available.
I reverted the scheme changes, sorry for changing them.
(I saw this tip a tweet from Jon Reid, if we detach the debugger from the scheme, the test execution speed will be much faster.)

MVP Module/Source/Feed Presentation/FeedLoadingViewModel.swift
```diff
@@ -4,4 +4,10 @@
 
 struct FeedLoadingViewModel {
 	let isLoading: Bool
+	let errorMessage: String?
```
caiozullo, line 7: The concept of `loading` and `error` should be separate since `errors` are not just for loading errors. We can represent errors for many other operations.

So ideally, there should be a separate model for the error, as shown in the lectures.

```swift
struct FeedErrorViewModel {
    let errorMessage: String?
}
```

Following the same idea, it should have a separate error view protocol too, to decouple the concepts of loading and error:

```swift
protocol FeedErrorView {
}
```

This way, you could have added this new feature without changing any code related to `loading`!

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -22,6 +23,7 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	}
 
 	@IBAction private func refresh() {
+		errorView?.hideMessage()
```
caiozullo, line 26: The Presenter should decide when to hide/show errors. See my other comment.

For example, in the `didStartLoadingFeed()` method, the Presenter can call  `errorView.display(FeedErrorViewModel(message: .none))` to hide the error.

This way, the ViewController doesn't need to decide when to hide the error. It's easier to manage presentation state transitions when all changes are centralized in the presenter.

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -9,6 +9,7 @@ protocol FeedViewControllerDelegate {
 }
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView {
+	@IBOutlet var errorView: ErrorView?
```
caiozullo, line 12: This property setter can be private to prevent clients from changing the instance:

```suggestion
	@IBOutlet private(set) var errorView: ErrorView?
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,15 @@ final class FeedViewModel {
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadFailure: Observer<String>?
```
caiozullo, line 23: `onFeedLoadFailure: Observer<String>?` could be `onErrorStateChange: Observer<String?>?`, so the ViewModel decides when to show/hide errors (instead of the view controller). It's easier to manage presentation state transitions when all changes are centralized.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,19 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+				self?.errorView?.hideMessage()
```
caiozullo, line 32: The ViewModel should decide when to hide/show errors. See my other comment.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -5,6 +5,8 @@
 import UIKit
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching {
+	@IBOutlet var errorView: ErrorView?
```
caiozullo, line 8: This property setter can be private to prevent clients from changing the instance:

```suggestion
	@IBOutlet private(set) var errorView: ErrorView?
```

MVVM Module/Source/Feed UI/Views/Feed.storyboard
```diff
@@ -14,19 +14,41 @@
                         <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
-                        <view key="tableHeaderView" contentMode="scaleToFill" id="clu-9B-M8g">
-                            <rect key="frame" x="0.0" y="0.0" width="414" height="16"/>
+                        <view key="tableHeaderView" contentMode="scaleToFill" id="GQ0-CT-NZN" customClass="ErrorView" customModule="MVC" customModuleProvider="target">
```
caiozullo, line 17: This storyboard is in the MVVM solution, so the `customModule` should be set to MVVM.

```suggestion
                        <view key="tableHeaderView" contentMode="scaleToFill" id="GQ0-CT-NZN" customClass="ErrorView" customModule="MVVM" customModuleProvider="target">
```

caiozullo (commented):
Hi @sanelz, you almost passed the challenge!

Thank you 🙂

caiozullo:
Hi @sanelz! Do you need any help with the PR comments?

sanelz:
Hi @caiozullo! 

Thank you for asking and for the detailed review, It was really helpful. I don't need any help with the PR comments, but I have some questions related to the challenge.

1. In the MVC module in the `viewDidLoad()` the `refreshController?.refresh()` method is called which is calling `errorView.hideMessage()`. It's hiding the error view even if the error view isn't showed. Is it OK to make this step even if it's not necessarily needed? Or could we also use something like `errorView.hideMessageIfNeeded()` where we would check if the view is already hidden, if so then we wouldn't need to hide it? I'm struggling a lot in general with similar situations. What would be the best practice in general?

2. In the MVC module, why did we decide to put the `errorView` in the `FeedRefreshViewController` and handle the logic there and not in the `FeedViewController` or maybe somewhere else? Does the `FeedRefreshViewController` has so two responsibilities (handling the refresh status and handling the error view status) or is this still one responsibility because its similar?

3. In the MVP module, we introduced a new `FeedErroView` protocol to handle the error view state changes. If we would have more views, we would end up with more protocols. After a while we would have lots of parameters in the presenter, how can we avoid this? Grouping some of the parameters together in an another class? What is the optimal number of the parameters? What would be here the best approach?
 
4. We introduced `@IBOutlet private(set) errorView: ErrorView?`. Here we used `private(set)` instead of only `private`, because we need a reference in the snapshot tests in the MVVM and MVP module? Is this a trade-off decision? Should we try to focus to go with `private` and figure out a different approach in the tests, e.g. 
``` swift
		let errorView = tableView.tableHeaderView as? ErrorView
		errorView?.show(message: errorMessage)
	}
	...
}
``` 
Would this a better approach in the `FeedUISnapshotTests`?

5. We implemented features from MVC to MVVM and from MVVMV to MVP in this order. Should we follow always this order if we would like to go from one pattern to another, or after a while all the patterns will become second nature and we can go immediately with e.g. MVVM or MVP and skip MVC? What is your recommendation?

(Sorry for having lot's of questions and for the late answer, I was very busy those days)

Thank you!

caiozullo (commented):
Hi @sanelz,

- All tests are passing ✅
- Added an extra language ✅

Great job!

---

Regarding your questions:

1) In the MVC solution, the controller is responsible for hiding the error (there's no presenter or view model to control the state transition). So every time we refresh, the controller tries to hide the error - if the error is already hidden, that's fine, it won't have any issues. You could check before performing the call, but that's not an expensive operation. Moreover, the initial error state may not be hidden (e.g., depending on how the storyboard is set). That's why on viewDidLoad, it may be important to reset it (hide it). But again, you could check if the error is already hidden before doing it if you prefer.

2) Because we wanted to show how the responsibilities can be split into multiple controllers, as we show in the lectures. It could also be all in one controller if the logic is simple enough. It's a choice. I recommend you rewatch the UI module lectures to learn more.

3) This is explained in the program. Follow the lectures in order, and you'll see how you can add many features without adding loads of protocols or properties in the presenter.

4) We recommend exposing the internal properties for testing purposes when needed. This way, the test doesn't need to know about the view hierarchy (e.g., the test doesn't need to know the error view is the table header view). So if we change the view hierarchy, we don't break the tests. If the test is coupled with the hierarchy, it's fragile to changes in production. If you change the structure, you'd have to update many tests. 

5) There's no order. You can follow any order when refactoring your code. Moreover, you don't need to transition from one to the other. Usually, you'll choose one pattern and use it throughout your codebase, so it's consistent.

Keep up the great work!

PR #99, opened 2021-06-26
Kenny Zheng - UI Design Patterns Challenge

firerainky:
Hi, I stuck on first test "test_emptyFeed" in `FeedUISnapshotTests` in MVVM module.

I cannot figure out what I am doing wrong.

And I have compared the storyboard layout in MVC module and the storyboard layout I implemented in MVVM module. 

The only difference I found is that in storyboard from MVC module, the button's content inset in bottom is 8
Screen Shot 2021-06-26 at 10 55 33 PM

and in storyboard I implemented from MVVM module, the button's content inset in bottom is 0.
Screen Shot 2021-06-26 at 10 58 02 PM

If I change the button's content inset in bottom to 8, the text of button will no longer be vertically center.

Can you explain why the button's content inset in bottom should be 8 to make it be vertically center in MVC module?

I tried to copy the `ErrorView` from MVC module's storyboard to MVVM module's storyboard, but the result is the same, there is still a white view below the error view in snapshot png (Dark mode)
Screen Shot 2021-06-26 at 11 12 56 PM

I have ran the app, and in MVC tab, the view below the error view is black in dark mode, what is the view below error view and why this view is white in my MVVM implementation.

I will be very appreciated if you can answer all my questions.

Thank you


firerainky:
I figure out the view below error view is the table footer view, and change it's color will make the test passing.
But I still can not understand why button's content inset in bottom should be 8 to make it be vertically center in MVC module.

Can you help me in this question?

firerainky:
Hi, I get stuck again, the second test `test_feedWithError` in Snapshot Test from MVVM module fails since the height of error view from MVVM module is smaller than the height of error view from MVC module.

But I can not figure out why, the height of error view from MVVM module is 34 px, the same as error view from MVC module.

When I ran the app, the height of error view in MVC tab is obviously greater than the height of error view in MVVM tab.

The problem looks like related with my last question above.

Can you help me out? Tank you

firerainky:
Hi, I get stuck again, the second test `test_feedWithError` in Snapshot Test from MVVM module fails since the height of error view from MVVM module is smaller than the height of error view from MVC module.

But I can not figure out why, the height of error view from MVVM module is 34 px, the same as error view from MVC module.

When I ran the app, the height of error view in MVC tab is obviously greater than the height of error view in MVVM tab.

The problem looks like related with my last question above.

Can you help me out? Tank you

caiozullo:
Hi @firerainky. The MVP and MVVM UI must match the MVC snapshots exactly, including the position, exact color, and font values. Please, check the `ErrorView` setup (position, color, fonts, etc) in the MVC storyboard and replicate it precisely in the MVP and MVVM storyboards. Also, check and replicate the constraints exactly - maybe there's a wrong constraint pushing the button off the center.

Make sure all tests are passing locally before pushing (select the CI scheme and press CMD+U to run all tests). And make sure you're using the correct simulator - check the readme file for all the details.

Let us know once you're done so we can start the review. If you can't solve the problem, let me know too so I can help you further. Thank you!

caiozullo:
Hi @firerainky. It's been a while. Do you need any help sorting out the issues?

caiozullo:
Hi @firerainky. Do you need any help sorting out the issues? We're here to help.

PR #100, opened 2021-06-27
Alex Kostenko - UI Design Patterns Challenge.

oligazar:


mapostolakis:
Hi @oligazar, if your PR is ready for review, please post a comment on the challenge page in the academy with the link to your PR. Thank you 🙂

thiagopenna:
Hi @oligazar ,

This is Penna, I am an instructor at Essential Developer.

Please, undo all changes to the schemes (they shouldn't have been changed - please check the readme for more details). Let me know once you're done so I can start the review. Thank you! 🙂

MVVM Module/Tests/Feed UI/Helpers/UIControl+TestHelpers.swift
```diff
@@ -8,7 +8,7 @@ extension UIControl {
 	func simulate(event: UIControl.Event) {
 		allTargets.forEach { target in
 			actions(forTarget: target, forControlEvent: event)?.forEach {
-				(target as NSObject).perform(Selector($0))
+				(target as NSObject).perform(Selector($0), with: nil)
```
thiagopenna, line 11: There's no need to use `perform(_: with:)` instead of `perform(_:)`, since all actions performed are `IBActions` that have no parameters. We would use this method if when creating the actions we selected `Arguments: Sender`, and had a reason to differentiate the object.
```suggestion
				(target as NSObject).perform(Selector($0))
```
According to the [documentation](https://developer.apple.com/documentation/objectivec/nsobjectprotocol/1418764-perform) of `perform(_: with:)`:
> This method is the same as perform(_:) except that you can supply an argument for aSelector. aSelector should identify a method that takes a single argument of type id. For methods with other argument types and return values, use NSInvocation.

thiagopenna (commented):
Hi @oligazar, you almost passed the challenge!

Thank you 🙂

MVVM Module/Tests/Feed UI/Helpers/UIControl+TestHelpers.swift
oligazar: I don't have enough experience with Storyboards. So I created the action with default sender argument (which is by default). And this lead to `EXC_BAD_ACCESS` exception which I had a hard time debugging. I had no idea what was the reason (it's at this point I changed the scheme hoping to find out the reason). 

After figuring out the issue I decided to keep perform `improved`, as I thought that this can be useful when testing actions with arguments. And this makes me wonder:

I understand now why it's suggested to create actions without arguments if we don't use them. But why this small redundancy is not welcomed in `.perform(:)`, since this method can be useful when testing IBActions with arguments?

caiozullo: @oligazar there's nothing wrong with using `perform(:)` in your personal projects but this change in the test helper is not part of this challenge. To pass the challenge, none of the helpers should have been changed.

oligazar:
@thiagopenna, as well as Caio and Mike, thank you for a great support and detailed explanations! I applied all the suggestions, so hope I did all correctly this time)

Just two small questions. I applied some suggestions from Thiago here on GitHub, but it turned out not to be safe, as we don't have compiler here to highlight potential issues. Those commits ended up being in "red state". So should we tend to apply suggestions in IDE as a rule of thumb?

And is there a way to fix this kind of undesired "red state" commits?

MVVM Module/Tests/Feed UI/Helpers/UIControl+TestHelpers.swift
thiagopenna: Hey @oligazar, my reasoning behind recommending this revert (apart from not being needed in this challenge, as @caiozullo said) is as follow:

* We are calling `IBActions` with no parameters, so there's no reason to provide a `sender`.
* If we had a reason to provide a `sender`, in no case it should be `nil`: we would actually pass the expected sender.
* In that case, we should probably create a whole new `simulate(event:sender:)` function, and deal with `Selectors` accordingly.

Although `perform(_:with:)` seems like an improvement, it can introduce other problems in the future where we assumed it would help, and one of them is as follows:
Screen Shot 2021-06-29 at 16 20 20

If instead of declaring the `Sender` argument as type `Any`, another developer declared it as type `UIButton` (something quite possible to happen using storyboards), there's no way for him to know that in another part of code this method will be called using the Objective C runtime passing a `nil` value, and in that case where he should be trusting the Swift Type System, he cannot. (In the example image I got a crash in a supposedly unreachable line of code)

So my main problem with leaving the code as is, is that `nil` is not a valid value to be used. It might prevent some crashes but might introduce others in unrelated places. And in the end, it's trying to solve the wrong problem: we are not using the `sender` argument anywhere.

thiagopenna:
Hi @oligazar ,

- All tests are passing ✅
- Added an extra language ✅

Also, regarding your questions:
The comments here are suggestions, intended to guide you in improving your solution, but intentionally left incomplete. We recommend you implement them on Xcode to practice and make sure there are no warnings or build errors.

Also, there's the possibility on GitHub to add suggestions to a batch and commit all of them simultaneously. For instance, when one specific change (like a rename) would alter multiple lines/files, there might be many suggestions dealing with that change and all of them could be committed together.

In the case where the applied suggestions leave the system in a "red state", there's the possibility to perform an `amend` and fix the last commit, or perform a `rebase` to deal with previous commits. In both cases, you should be really careful since you would be rewriting git history.

Great job!

PR #108, opened 2021-07-27
Joe Lauletta - UI Design Patterns Challenge

jlappdev:


MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,12 +13,20 @@ protocol FeedView {
 	func display(_ viewModel: FeedViewModel)
 }
 
+protocol FeedErrorView {
+	typealias ErrorMessage = String
+
+	func display(_ errorMessage: ErrorMessage)
+}
```
danillahtin, line 20: The `message` could be an optional `String?`, so the Presenter can decide when to show/hide the message.

For example, in the `didStartLoadingFeed()` method, the Presenter can call  `feedErrorView.display(.none)` to hide the error.

This way, the ViewController doesn't need to decide when to hide the error. It's easier to manage presentation state transitions when all changes are centralized in the presenter.

Also It can be a bit more convenient to have a `FeedErrorViewModel` struct to describe error view instead of a `typealias ErrorMessage = String`. It is more typesafe and it would be easier to add new parameter (say some error icon) to `FeedErrorView` in future as it would require only the change in `FeedErrorViewModel` and will not require change in `FeedErrorView` protocol.

```suggestion
struct FeedErrorViewModel {
	let message: String?
}

protocol FeedErrorView {
	func display(_ viewModel: FeedErrorViewModel)
}
```

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -22,6 +24,7 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 	}
 
 	@IBAction private func refresh() {
+		errorView?.hideMessage()
```
danillahtin, line 27: The Presenter should decide when to hide/show errors. See my other comment.

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +21,15 @@ final class FeedViewModel {
...
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onErrorThrown: ErrorObserver?
 
 	func loadFeed() {
 		onLoadingStateChange?(true)
 		feedLoader.load { [weak self] result in
 			if let feed = try? result.get() {
 				self?.onFeedLoad?(feed)
+			} else {
+				self?.onErrorThrown?()
```
danillahtin, line 32: `onErrorThrown: ErrorObserver?` could be `onErrorStateChange: Observer<String?>?`, so the ViewModel decides when to show/hide errors (instead of the view controller) and which text to present. It's easier to manage presentation state transitions when all changes are centralized.

```suggestion
	var onErrorStateChange: Observer<String?>?

	func loadFeed() {
		onLoadingStateChange?(true)
		onErrorStateChange?(nil)
		feedLoader.load { [weak self] result in
			if let feed = try? result.get() {
				self?.onFeedLoad?(feed)
			} else {
				self?.onErrorStateChange?(Localized.Feed.loadError)
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -21,6 +23,7 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 
 	@IBAction private func refresh() {
 		viewModel?.loadFeed()
+		errorView?.hideMessage()
```
danillahtin, line 26: The ViewModel should decide when to hide/show errors. See my other comment.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -32,6 +35,10 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 				self?.refreshControl?.endRefreshing()
 			}
 		}
+
+		viewModel?.onErrorThrown = { [weak self] in
+			self?.errorView?.show(message: Localized.Feed.loadError)
```
danillahtin, line 40: The ViewModel should decide which error text to display. See my other comment.

danillahtin (commented):
Hi @jlappdev, you almost passed the challenge!

Thank you 🙂

MVVM Module/Tests/Feed UI/FeedUISnapshotTests.swift
```diff
@@ -65,7 +65,7 @@ private class FeedLoaderStub: FeedLoader {
 
 private extension FeedViewController {
 	func display(errorMessage: String) {
-		fatalError("Must be implemented - follow the MVC solution as a guide")
+		errorView?.show(message: errorMessage)
```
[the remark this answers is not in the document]
jlappdev: @danillahtin I wanted to better understand the distinct between using the `FeedViewController` reference to errorView and its reference to viewModel. Is the viewModel property not considered an internal property of `FeedViewController`?

jlappdev:
@danillahtin I have addressed all of your concerns. Please review the changes. Also, I did have one question for you concerning a comment above. Thanks!

MVVM Module/Tests/Feed UI/FeedUISnapshotTests.swift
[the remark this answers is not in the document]
danillahtin: Ideally `FeedViewController.viewModel` should not be used as well. It is better to control the viewModel from the outside. But due to the tests configuration and lack of external reference to the viewModel we can choose the lesser evil and address the viewModel via `FeedViewController.viewModel`.

Ideal solution would be to create new viewModel and then control it:
```swift
        let viewModel = FeedViewModel(feedLoader: FeedLoaderStub(.success([])))
        self.viewModel = viewModel

        viewModel.onErrorStateChange?(errorMessage)
}
```
But as you can see it takes 3 times more code to implement than not ideal one.

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -9,6 +9,8 @@ protocol FeedViewControllerDelegate {
 }
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView {
+	@IBOutlet private var errorView: ErrorView?
```

```suggestion
	@IBOutlet public private(set) var errorView: ErrorView?
```

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -57,4 +57,16 @@ extension FeedViewController {
 	private var feedImagesSection: Int {
 		return 0
 	}
+
+	var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 64: To follow up on my previous comment this getter can be removed

```suggestion
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -5,6 +5,8 @@
 import UIKit
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching {
+	@IBOutlet private var errorView: ErrorView?
```
danillahtin, line 8: Following the same logic as for MVP solution I recommend making `errorView` getter public and setter private.

```suggestion
	@IBOutlet public private(set) var errorView: ErrorView?
```

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -57,4 +57,16 @@ extension FeedViewController {
 	private var feedImagesSection: Int {
 		return 0
 	}
+
+	var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 64: To follow up on my previous comment this getter can be removed
```suggestion
```

danillahtin (commented):
Hi @jlappdev, almost there!

Thank you 🙂

jlappdev:
@danillahtin I have reverted the previous commits I had made which made the errorView property private. Thanks!

danillahtin (commented):
Hi @jlappdev,

- All tests are passing ✅
- Added an extra language ✅

Great job!

PR #109, opened 2021-07-27
Caio Ortu - UI Design Patterns Challenge.

caioortu:


MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -8,7 +8,9 @@ protocol FeedViewControllerDelegate {
 	func didRequestFeedRefresh()
 }
 
-public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView {
+public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView, FeedErrorView {
+	@IBOutlet var errorView: ErrorView?
```

```suggestion
	@IBOutlet public private(set) var errorView: ErrorView?
```

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -40,6 +40,18 @@ extension FeedViewController {
 		ds?.tableView?(tableView, cancelPrefetchingForRowsAt: [index])
 	}
 
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 46: To follow up on my previous comment this getter can be removed

```suggestion
```

MVP Module/Tests/Feed UI/FeedUISnapshotTests.swift
```diff
@@ -51,6 +51,6 @@ class FeedUISnapshotTests: XCTestCase {
 
 private extension FeedViewController {
 	func display(errorMessage: String) {
-		fatalError("Must be implemented - follow the MVC solution as a guide")
+		errorView?.show(message: errorMessage)
```
danillahtin, line 54: If at some point error displaying logic will get more complex other than just simply showing a message in `errorView` then this method would potentially break the test. To decouple the test I recommend using `FeedErrorViewModel` and `display` method of `FeedErrorView` protocol.
```suggestion
		display(FeedErrorViewModel(message: errorMessage))
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,16 @@ final class FeedViewModel {
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadError: Observer<String>?
```
danillahtin, line 23: `onFeedLoadError: Observer<String>?` could be `onErrorStateChange: Observer<String?>?`, so the ViewModel decides when to show/hide errors (instead of the view controller) and naming is consistent with `onLoadingStateChange`. It's easier to manage presentation state transitions when all changes are centralized.

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -5,6 +5,8 @@
 import UIKit
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching {
+	@IBOutlet var errorView: ErrorView?
```
danillahtin, line 8: The same as for the MVP solution, I recommend `errorView` to have a public getter and a private setter

```suggestion
	@IBOutlet public private(set) var errorView: ErrorView?
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,16 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+				self?.errorView?.hideMessage()
```
```suggestion
```

MVVM Module/Tests/Feed UI/FeedUISnapshotTests.swift
```diff
@@ -65,7 +65,7 @@ private class FeedLoaderStub: FeedLoader {
 
 private extension FeedViewController {
 	func display(errorMessage: String) {
-		fatalError("Must be implemented - follow the MVC solution as a guide")
+		errorView?.show(message: errorMessage)
```
danillahtin, line 68: For the same logic as for the MVP solution I recommend to display error message using the viewModel here:

```suggestion
		viewModel?.onErrorStateChange?(errorMessage)
```

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -40,6 +40,18 @@ extension FeedViewController {
 		ds?.tableView?(tableView, cancelPrefetchingForRowsAt: [index])
 	}
 
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 46: To follow up on my previous comment this getter can be removed

```suggestion
```

danillahtin (commented):
Hi @caioortu, you almost passed the challenge!

Thank you 🙂

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
caioortu: I did this change, but this kept me wondering because it seems weird to change the error view visibility just for the tests, also the MVC solution uses the same logic so maybe changes should be made there too

danillahtin: Hi @caioortu,
You are right that this is done for tests only and this solution is not ideal. On the other hand it helps to decouple the tests from the view hierarchy in the storyboard.   Programming is all about trade offs. If you and your team

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -37,6 +39,14 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		tableModel = cellControllers
 	}
 
+	func display(_ viewModel: FeedErrorViewModel) {
+		guard let message = viewModel.message else {
+			errorView?.hideMessage()
+			return
+		}
+		errorView?.show(message: message)
```
danillahtin, line 47: Suggestion: 

An `if/else` statement is usually easier to read/follow in such cases:

"If there's a message to show, show it. Else, hide the existing one."

```suggestion
		if let message = viewModel.message {
			errorView?.show(message: message)
		} else {
			errorView?.hideMessage()
		}
```

Instead of "guard against not having a message. Else, hide the existing one. Show the message if it passed the guard statement."

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,17 @@ final class FeedViewModel {
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadError: Observer<String?>?
```
danillahtin, line 23: `onFeedLoadError` could be named `onErrorStateChange` following the `onLoadingStateChange` property.
```suggestion
	var onErrorStateChange: Observer<String?>?
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
danillahtin: Now as viewModel controls error view visibility, this line can be removed. Please remove it

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,20 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
 				self?.refreshControl?.endRefreshing()
 			}
 		}
+
+		viewModel?.onFeedLoadError = { [weak self] errorMessage in
+			guard let message = errorMessage else {
+				self?.errorView?.hideMessage()
+				return
+			}
+			self?.errorView?.show(message: message)
```
danillahtin, line 44: Same suggestion here - an `if/else` statement can be easier to read/follow.

danillahtin (commented):
Hi @caioortu, almost there!

Thank you 🙂

PR #117, opened 2021-08-21
Haik Ampardjian - UI Design Patterns Challenge

ludoded:


MVP Module/Source/Feed Presentation/FeedErrorViewModel.swift
```diff
@@ -0,0 +1,11 @@
+//
+//  FeedErrorViewModel.swift
+//  MVP
+//
+//  Created by Haik Ampardjian on 21.08.2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+struct FeedErrorViewModel {
+	let errorMessage: String
```
danillahtin, line 10: Suggestion: 
Since it is an **error** viewModel, the context is pretty clear and the property can be named simply `message`.
Also the message can be optional following my next comment on `FeedErrorView`
```suggestion
	let message: String?
```

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,17 +13,25 @@ protocol FeedView {
 	func display(_ viewModel: FeedViewModel)
 }
 
+protocol FeedErrorView {
+	func display(_ viewModel: FeedErrorViewModel)
+	func hide()
+}
```
danillahtin, line 19: Suggestion:
It can be better to have `FeedErrorViewModel.errorMessage` as optional and hide the view by calling `display(.init(errorMessage: nil))`, because `hide` is a very generic name and can have conflicts with other methods in the future. 
Also it will keep all the state transitions centralized into a single state type.

```suggestion
protocol FeedErrorView {
	func display(_ viewModel: FeedErrorViewModel)
}
```

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -8,7 +8,9 @@ protocol FeedViewControllerDelegate {
 	func didRequestFeedRefresh()
 }
 
-public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView {
+public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView, FeedErrorView {
+	@IBOutlet private(set) weak var errorView: ErrorView!
```

```suggestion
	@IBOutlet public private(set) weak var errorView: ErrorView!
```

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -40,6 +40,18 @@ extension FeedViewController {
 		ds?.tableView?(tableView, cancelPrefetchingForRowsAt: [index])
 	}
 
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 46: As a follow up on my previous comment, this getter can be removed.
```suggestion
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -5,6 +5,8 @@
 import UIKit
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching {
+	@IBOutlet private(set) weak var errorView: ErrorView!
```
danillahtin, line 8: Same here as for the MVP solution:
```suggestion
	@IBOutlet public private(set) weak var errorView: ErrorView!
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,23 @@ final class FeedViewModel {
 		Localized.Feed.title
 	}
 
+	var errorMessage: String {
+		Localized.Feed.feedLoadingError
+	}
+
```
danillahtin, line 24: It is better not to have `errorMessage` as a getter is that at some point error message can be decided to be sent from server instead of being bundled with the app. Also it can be more errors and thus more different messages.
Thus it is better to send the message via `onErrorStateChange` block.
```suggestion
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,23 @@ final class FeedViewModel {
...
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onErrorStateChange: Observer<Bool>?
```
danillahtin, line 27: As a follow up on my previous comment, this block can be `Observer<String?>?`

```suggestion
	var onErrorStateChange: Observer<String?>?
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,23 @@ final class FeedViewModel {
...
 
 	func loadFeed() {
 		onLoadingStateChange?(true)
+		onErrorStateChange?(false)
```
danillahtin, line 31: As a follow up on my previous comments, nil can be sent here instead of `false` 
```suggestion
		onErrorStateChange?(nil)
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -18,14 +18,23 @@ final class FeedViewModel {
...
 		feedLoader.load { [weak self] result in
-			if let feed = try? result.get() {
+			switch result {
+			case let .success(feed):
 				self?.onFeedLoad?(feed)
+			case .failure:
+				self?.onErrorStateChange?(true)
```
danillahtin, line 37: As a follow up on my previous comments, here the exact error message should be sent.
```suggestion
				self?.onErrorStateChange?(Localized.Feed.feedLoadingError)
```

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -40,6 +40,18 @@ extension FeedViewController {
 		ds?.tableView?(tableView, cancelPrefetchingForRowsAt: [index])
 	}
 
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 46: The same here as for the MVP solution, the getter can be removed:
```suggestion
```

danillahtin (commented):
Hi @ludoded, you almost passed the challenge!

Thank you 🙂

MVP Module/Source/Feed Presentation/FeedPresenter.swift
ludoded: @danillahtin Thanks for your comments!

I was actually thinking about that approach.
But what got me off the track with that solution is that I have to make the message of an optional String type.
And the problem with that is it's not explicit. 

For instance, for me to read this code, and see `hide` and `display` methods are more self-documented and explicit than having a single `display` method that can either hide or display.
Single method `display` that can also hide, in my opinion, is a violation of SoC principle. 

To not cause any overriding issues in the future and conform to proper polymorphism rules, I can rename the `hide` method to `hideErrorView`.

Let me know what you think, and if you still want me to make those changes, I'll do them gladly.

Thanks!

ludoded:
@danillahtin 
Thanks for your review!
Updates are pushed.

MVP Module/Source/Feed Presentation/FeedPresenter.swift
danillahtin: Hi @ludoed,
All good, it was just a suggestion, either ways are correct 👍

danillahtin (commented):
Hi @ludoded,

Congratulations, you've passed the first challenge and got your first belt stripe ⚪️①!

✅ All tests are passing, so your implementation complies with the specs.

**[Update your status](https://iosleadessentials.slack.com/archives/CTAR0LVMM/p1629389971188200) in the community now with the following emoji to celebrate your achievement: `:white-belt-1:`**

Great job, and keep it up! Remember, we're here to help you progress and get the most out of this program.

See you in the community and in the following challenge 🙂

PR #128, opened 2021-09-05
Luis Zapata - UI Design Patterns Challenge

zgluis:


MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -57,4 +57,16 @@ extension FeedViewController {
 	private var feedImagesSection: Int {
 		return 0
 	}
+
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
```
danillahtin, line 63: Suggestion:
You can make `FeedErrorViewController.errorView` getter public on the production side and there will be no need to introduce this getter here.
```suggestion
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -13,6 +13,8 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		didSet { tableView.reloadData() }
 	}
 
+	@IBOutlet weak var errorView: ErrorView!
```
danillahtin, line 16: Same here for as for the MVP solution:
```suggestion
	@IBOutlet private(set) var errorView: ErrorView!
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,16 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+				self?.errorView.hideMessage()
```
danillahtin, line 32: It is not the controller but the viewModel who should hide the message. In order to do that, `onFeedLoadFailure: Observer<String>?` type can be changed to `Observer<String?>?` and in case there's a `nil` message then the view should be hidden.
Please move this logic to the viewModel
```suggestion
```

MVVM Module/Source/Feed UI/Views/Feed.storyboard
```diff
@@ -14,19 +14,41 @@
                         <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
-                        <view key="tableHeaderView" contentMode="scaleToFill" id="clu-9B-M8g">
-                            <rect key="frame" x="0.0" y="0.0" width="414" height="16"/>
+                        <view key="tableHeaderView" contentMode="scaleToFill" id="clu-9B-M8g" userLabel="Error VIew" customClass="ErrorView" customModule="MVVM" customModuleProvider="target">
```
danillahtin, line 17: There's a typo in userLabel:
```suggestion
                        <view key="tableHeaderView" contentMode="scaleToFill" id="clu-9B-M8g" userLabel="Error View" customClass="ErrorView" customModule="MVVM" customModuleProvider="target">
```

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -57,4 +57,16 @@ extension FeedViewController {
 	private var feedImagesSection: Int {
 		return 0
 	}
+
+	private var errorView: ErrorView? {
+		tableView.tableHeaderView as? ErrorView
+	}
+
```
danillahtin, line 64: Suggestion:
Same here as for the MVP solution, getter can be removed if `errorView` getter is public on the production side:
```suggestion
```

danillahtin (commented):
Hi @zgluis, you almost passed the challenge!

Thank you 🙂

danillahtin:
Hi @zgluis. Do you need any help performing the changes requested in the review? Just let me know - We're here to help! 🙂

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
zgluis: Hi @danillahtin , thanks for the suggestion, I'm applying the change but, could you please explain more about what is the benefit of taking this approach?

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
zgluis: Hi @danillahtin thanks for the comment. I'm not sure if I get it, how would the viewModel call `errorView.hideMessage()` if it doesn't have a reference to the `errorView`?

zgluis:
> Hi @zgluis. Do you need any help performing the changes requested in the review? Just let me know - We're here to help! 🙂

Hey @danillahtin , sorry about the delay. I addressed most of your comments please your help reviewing the changes, thanks!

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
danillahtin: Hi @zgluis, The only benefit here is less code

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +21,15 @@ final class FeedViewModel {
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadFailure: Observer<ErrorMessage>?
```
danillahtin, line 24: `onFeedLoadFailure: Observer<ErrorMessage>?` could be `onFeedLoadFailure: Observer<ErrorMessage?>?`, so the ViewModel decides when to show/hide errors (instead of the view controller). It's easier to manage presentation state transitions when all changes are centralized.

For example, in the `loadFeed()` method, the ViewModel can call  `onFeedLoadFailure(.none)` to hide the error before loading the feed.

```suggestion
	var onFeedLoadFailure: Observer<ErrorMessage?>?
```

Suggestion:
`ErrorMessage` typealias makes the code less readable since every time you see a new id/name in the codebase you have to go and figure out what it stands for, it can be very confusing. So it is better to keep things simple and use plain `String`.

Another suggestion is to name this observer as `onErrorStateChange`. Error can happen not only during load but by some other reason. Also this name is consistent with `onLoadingStateChange` which make the code a little bit more readable.
```suggestion
	var onErrorStateChange: Observer<String?>?
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
danillahtin: Hi @zgluis,
I answered this question in the other comment (it's on `FeedViewModel`), please remove this line

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -13,6 +13,8 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		didSet { tableView.reloadData() }
 	}
 
+	@IBOutlet public var errorView: ErrorView!
```
danillahtin, line 16: The setter can be private to prevent clients from changing the instance:

```suggestion
	@IBOutlet public private(set) var errorView: ErrorView!
```

MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,16 +13,24 @@ protocol FeedView {
 	func display(_ viewModel: FeedViewModel)
 }
 
+protocol FeedErrorView {
+	func show(message: String)
+	func hide()
+}
+
```
danillahtin, line 20: Suggestion:
`func show(message: String)` can have optional `message` (`show(message: String?)`). The hiding can be done by calling `show(message: nil)`, because `hide` is a very generic name and can have conflicts with other methods in the future. 
Also it will keep all the state transitions centralized into a single state type (`String?`).

For a single state type It can be better to have a separate `FeedErrorViewModel` that describes the error view state (the same way that `FeedLoadingViewModel` does). It will make introducing new parameters a lot easier. With current approach I will need to modify `FeedErrorView.show` function which will make `FeedPresenter` and `FeedViewController` stop compiling. 

```suggestion
struct FeedErrorViewModel {
	let message: String?
}

protocol FeedErrorView {
	func display(_ viewModel: FeedErrorViewModel)
}
```

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -21,6 +21,8 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		refresh()
 	}
 
+	@IBOutlet public var errorView: ErrorView!
```
danillahtin, line 24: The setter can be private to prevent clients from changing the instance:

```suggestion
	@IBOutlet public private(set) var errorView: ErrorView!
```

danillahtin (commented):
Hi @zgluis, almost there!

Thank you 🙂

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
zgluis: Hey @danillahtin thanks, I added the suggested change

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
zgluis: Thanks!

zgluis:
Hey @danillahtin thanks for the comments, I applied the changes 🙌

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,20 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+				self?.errorView.hideMessage()
```
danillahtin, line 32: The viewModel already hides the `errorView` when load starts, no need to do it once again here
Please remove the line
```suggestion
```

danillahtin (commented):
Hi @zgluis, almost there!

Please, check my review comment and let me know once you're done so I can review it again. 

Thank you 🙂

PR #131, opened 2021-09-12
Lívia Vasconcelos  - UI Design Patterns Challenge.

liuvasconcelos:
Here is my PR for the UI Design Patterns Challenge.
Thanks for the review and any improvement suggestion.


MVP Module/Source/Feed Presentation/FeedPresenter.swift
```diff
@@ -13,16 +13,23 @@ protocol FeedView {
...
 }
 
+protocol FeedErrorView {
+	func display(_ viewModel: FeedErrorViewModel)
+}
+
 final class FeedPresenter {
 	private let feedView: FeedView
 	private let loadingView: FeedLoadingView
+	private let feedErrorView: FeedErrorView
```
danillahtin, line 23: Suggestion:
`feedErrorView` could be named just `errorView` to simplify - the context is clear here so we can avoid repeating "feed" many times (like the `loadingView` is not called `feedLoadingView`).

```suggestion
	private let errorView: FeedErrorView
```

If you follow this suggestion, please, rename it everywhere in this class.

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -8,7 +8,9 @@ protocol FeedViewControllerDelegate {
 	func didRequestFeedRefresh()
 }
 
-public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView {
+public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching, FeedLoadingView, FeedErrorView {
+	@IBOutlet weak var errorView: ErrorView!
```
danillahtin, line 12: This property doesn't need to be `weak`. Also, the setter can be private to prevent clients from changing the instance:

```suggestion
	@IBOutlet private(set) var errorView: ErrorView!
```

MVP Module/Source/Feed UI/Views/Feed.storyboard
```diff
@@ -14,19 +14,41 @@
                         <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
-                        <view key="tableHeaderView" contentMode="scaleToFill" id="clu-9B-M8g">
-                            <rect key="frame" x="0.0" y="0.0" width="414" height="16"/>
+                        <view key="tableHeaderView" contentMode="scaleToFill" id="znp-2z-PVb" customClass="ErrorView" customModule="MVC" customModuleProvider="target">
```
danillahtin, line 17: This storyboard is in the MVP solution, so the `customModule` should be set to MVP.

```suggestion
                        <view key="tableHeaderView" contentMode="scaleToFill" id="znp-2z-PVb" customClass="ErrorView" customModule="MVP" customModuleProvider="target">
```

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -57,4 +57,16 @@ extension FeedViewController {
 	private var feedImagesSection: Int {
 		return 0
 	}
+
+	var errorView: ErrorView? {
```
danillahtin, line 61: This helper is not used from outside of this file thus it can be private, please make it private
```suggestion
	private var errorView: ErrorView? {
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -5,6 +5,8 @@
 import UIKit
 
 public final class FeedViewController: UITableViewController, UITableViewDataSourcePrefetching {
+	@IBOutlet weak var errorView: ErrorView!
```
danillahtin, line 8: This property doesn't need to be `weak`. Also, the setter can be private to prevent clients from changing the instance:

```suggestion
	@IBOutlet private(set) var errorView: ErrorView!
```

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,13 +20,18 @@ final class FeedViewModel {
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onFeedLoadError: Observer<Error>?
```
danillahtin, line 23: `Error` is a model type. The view that subscribes to `onFeedLoadError` should then decide how to display this error. This decision relates to Presentation logic, not to UI. Thus it is the viewModel who should transform an `Error` into some view-displayable model. 
In this case the view-displayable model can be a simple `String?`. If the value is nil then we hide the errorView, if the value is not nil we show it with the message received. 

Putting it all together:
`onFeedLoadError: Observer<Error>?` could be `onFeedLoadError: Observer<String?>?`, so the ViewModel decides when to show/hide errors (instead of the view controller). It's easier to manage presentation state transitions when all changes are centralized.

For example, in the `loadFeed()` method, the ViewModel can call  `onFeedLoadError(.none)` to hide the error before loading the feed.

```suggestion
	var onFeedLoadError: Observer<String?>?
```

Suggestion:
Since error can emerge not only because of loading but for some other reasons it would be more scalable not to couple error displaying with loading and call the observer `onErrorStateChange` instead. Thus it would be more scalable and consistent with `onLoadingStateChange`.
```suggestion
	var onErrorStateChange: Observer<String?>?
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,15 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+				self?.errorView.hideMessage()
```
```suggestion
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,15 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
...
 				self?.refreshControl?.beginRefreshing()
 			} else {
 				self?.refreshControl?.endRefreshing()
 			}
 		}
+		viewModel?.onFeedLoadError = { [weak self] _ in
+			self?.errorView.show(message: Localized.Feed.loadError)
+		}
```
danillahtin, line 40: The ViewModel should decide which message to show. See my other comment.
```suggestion
		viewModel?.onFeedLoadError = { [weak self] message in
			if let message = message {
				self?.errorView.show(message: message)	
			} else {
				self?.errorView.hideMessage()
			}
		}
```

MVVM Module/Source/Feed UI/Views/Feed.storyboard
```diff
@@ -14,19 +14,41 @@
                         <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
-                        <view key="tableHeaderView" contentMode="scaleToFill" id="clu-9B-M8g">
-                            <rect key="frame" x="0.0" y="0.0" width="414" height="16"/>
+                        <view key="tableHeaderView" contentMode="scaleToFill" id="KcC-1v-niI" customClass="ErrorView" customModule="MVC" customModuleProvider="target">
```
danillahtin, line 17: This storyboard is in the MVP solution, so the `customModule` should be set to MVP.

```suggestion
                        <view key="tableHeaderView" contentMode="scaleToFill" id="KcC-1v-niI" customClass="ErrorView" customModule="MVVM" customModuleProvider="target">
```

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
```diff
@@ -57,4 +57,16 @@ extension FeedViewController {
 	private var feedImagesSection: Int {
 		return 0
 	}
+
+	var errorView: ErrorView? {
```
danillahtin, line 61: This helper is not used from outside of this file thus it can be private, please make it private
```suggestion
	private var errorView: ErrorView? {
```

danillahtin (commented):
Hi @liuvasconcelos, you almost passed the challenge!

Thank you 🙂

MVP Module/Source/Feed Presentation/FeedPresenter.swift
liuvasconcelos: Great! Done!

MVP Module/Source/Feed UI/Controllers/FeedViewController.swift
liuvasconcelos: Nice! Done

MVP Module/Source/Feed UI/Views/Feed.storyboard
liuvasconcelos: Ok, my mistake!

MVP Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
liuvasconcelos: Ok!

MVVM Module/Tests/Feed UI/Helpers/FeedViewController+TestHelpers.swift
liuvasconcelos: Ok!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
liuvasconcelos: Done

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
liuvasconcelos: Thanks! That was a really nice explanation!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
liuvasconcelos: Ok!

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
liuvasconcelos: Ok!

MVVM Module/Source/Feed UI/Views/Feed.storyboard
liuvasconcelos: Ok

liuvasconcelos:
Hi @danillahtin ! Thanks for the great review! 
I just pushed all modifications. Could you review again? 

Thanks!

PR #134, opened 2021-09-14
Anton Ihnatovich - UI Design Patterns Challenge

antonihnatovich:


MVVM Module/Source/Feed Presentation/FeedViewModel.swift
```diff
@@ -20,12 +20,17 @@ final class FeedViewModel {
 
 	var onLoadingStateChange: Observer<Bool>?
 	var onFeedLoad: Observer<[FeedImage]>?
+	var onError: Observer<String>?
```
danillahtin, line 23: `onError: Observer<String>?` could be `onErrorStateChange: Observer<String?>?`, so the ViewModel decides when to show/hide errors (instead of the view controller). It's easier to manage presentation state transitions when all changes are centralized.

```suggestion
	var onErrorStateChange: Observer<String?>?
```

MVVM Module/Source/Feed UI/Controllers/FeedViewController.swift
```diff
@@ -27,11 +29,16 @@ public final class FeedViewController: UITableViewController, UITableViewDataSou
 		title = viewModel?.title
 		viewModel?.onLoadingStateChange = { [weak self] isLoading in
 			if isLoading {
+				self?.errorView?.hideMessage()
```
```suggestion
```

danillahtin (commented):
Hi @antonihnatovich, you almost passed the challenge!

Thank you 🙂

MVVM Module/Source/Feed Presentation/FeedViewModel.swift
antonihnatovich: Hello @danillahtin 
I agree, it's better to name it like you proposed. Thanks for a clear explanation :)

